PR #1, opened 2017-12-17
Fix compile errors with Microsoft Visual Studio and STL

sercxjo:


sercxjo:
But for me VS 2015 produces an error about an unknown function inet_ntop

eichelberg:
That is quite weird - our nightly build includes Visual Studio 2015, both 32 and 64 bit, and it runs just fine (just did so last night, definitely including ofsockad.cc). Anyway, your patch should not do any harm. I will commit it and see if it causes problems with any other Win32 build; if it doesn't, it will stay in.

sercxjo:
Thanks, I want to suggest another patch for storescp (and dul) to work with national symbols in storing directory in argument list on Windows

eichelberg:
Apparently the inet_ntop() function is only available from Windows 7, so whether or not this compiles probably depends on the version of the Platform SDK you have installed with your Visual Studio installation. Anyway, your compile fix has meanwhile been committed since it also works that way.

With regard to the storescp patch: Please do. Support for non-ASCII file and directory names is indeed only partial in DCMTK. On Linux/Posix, this is a non-issue because these platforms use UTF-8, but on Windows it does require special handling.

PR #3, opened 2018-02-07
fixes for poll use

sercxjo:
The patch fixes errors with network connection after commit 5fae4f233a41232bca18595dd8b5c2515617855f under Linux.

### DIFF
--- a/dcmnet/libsrc/dcmtrans.cc
+++ b/dcmnet/libsrc/dcmtrans.cc
@@ -54,6 +54,7 @@ END_EXTERN_C
 
 #ifdef DCMTK_HAVE_POLL
 #include <poll.h>
+#include "dcmtk/ofstd/ofvector.h"
 #endif
 
 /* platform independent definition of EINTR */
@@ -192,7 +193,9 @@ OFBool DcmTransportConnection::fastSelectReadableAssociation(DcmTransportConnect
   SOCKET maxsocketfd = INVALID_SOCKET;
 #else
   int socketfd = -1;
+#ifndef DCMTK_HAVE_POLL
   int maxsocketfd = -1;
+#endif
 #endif
 
   int i=0;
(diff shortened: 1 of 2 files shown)


PavelTrushkin:
grrr, i spent 4 hours catching POLLOUT...

michaelonken:
@sercxjo Thanks for the report, do you have a test program that demonstrates the issue? Switching to POLLOUT when requesting an association sounds logical to me, but DCMTK network tests surprisingly still work on my linux box without applying your patch.

PavelTrushkin:
Michael, yes it works LOCALLY but will fail if  connect return EINPROGRESS 

Please check man poll there is explicitly required pillout after connect

PavelTrushkin:
In other words if you call poll(POLLIN) after connect/EINPROGRESS you will get timeout, I can reproduce it 100%

sercxjo:
We use different PACS servers (Orthanc and other) in other buildings with internet connection and our program based on modified storescp/storescu. Set all timeouts to 60s. All worked before 5fae4f233a41232bca18595dd8b5c2515617855f

michaelonken:
I looked it up on the connect() man page, and took over the POLLOUT versus the POLLIN. It will show up on DCMTK master during the next day. Thanks Sergej!

I'm now looking at the rest of your patch ;)

michaelonken:
Sergej, the POLLIN/POLLOUT part of the [patch is in DCMTK now for some time](http://git.dcmtk.org/?p=dcmtk.git;a=commit;h=0630c52fff6d0422a7aeadde521133fa888bf57f).

What does the rest of the patch do, i.e. the changes to dcmnet/libsrc/dcmtrans.cc?

eichelberg:
This patch has now been fully merged into DCMTK with commit #73107ff (2018-06-12).
Note that the patch contained one bug which has been fixed in the commit: The second poll() parameter must be "pfd.size()" and not "connCount".

michaelonken:
Thanks for taking care of this Marco

sercxjo:
Thanks

PavelTrushkin:
could you push commit 73107ff ? i don't see it.

michaelonken:
Just did that (every commit goes through a testing dashboard, that's why it can take a few days until we push it to the public branches).

PR #7, opened 2018-10-16
Fix PACKAGE_PATH in exported CMake package config

FlorianFranzen:
DCMTK_INSTALL_CMKDIR already contains the install prefix.

As a result, when the PACKAGE_PATH is computed in relation to the package config install path, the result is incorrect.

After installation ```find_package(dcmtk)``` therefore fails at the ```set_and_check(...)``` inside the generated ```DCMTKConfig.cmake``` as a result.


krop:
+1 We applied the same patch a couple months ago in openSUSE.

krop:
@jriesmeier @michaelonken

jriesmeier:
@krop I already forwarded the "pull request" to the DCMTK team member(s) who introduced this bug.

jriesmeier:
Should have been fixed by [this commit](https://github.com/DCMTK/dcmtk/commit/09255c29e40eca20ff9d06ca2e4f8906ffb47a9b).

PR #8, opened 2018-12-04
Add UserDataN names to private.dic.

eborisch:
[Reference](https://www.gehealthcare.com/-/media/1587e4b3d7b440e8b4d4178da553e085.pdf)

Just makes it a little nicer when looking for a specific one.


eichelberg:
This patch been fully merged into DCMTK today with commit #f3541261d.
The modification should appear in the public git repository and the
Github mirror in a couple of days.

PR #9, opened 2018-12-07
Require PkgConfig if you need to use it

agravgaard:
Because I suddenly got the following error in my CI:
`CMake Error at build/DCMTK-src/CMake/FindOpenJPEG.cmake:10 (pkg_check_modules):`
`  Unknown CMake command "pkg_check_modules".`
`Call Stack (most recent call first):`
`  build/DCMTK-src/CMake/3rdparty.cmake:340 (find_package)`
`  build/DCMTK-src/CMake/dcmtkPrepare.cmake:429 (include)`
`  build/DCMTK-src/CMakeLists.txt:17 (include)`

### DIFF
--- a/CMake/FindOpenJPEG.cmake
+++ b/CMake/FindOpenJPEG.cmake
@@ -7,6 +7,7 @@
 #  OPENJPEG_FOUND        - True if OpenJPEG found
 
 # Use pkg_check_modules to determine the paths for OpenJPEG
+find_package(PkgConfig REQUIRED)
 pkg_check_modules(OPENJPEG_PKGCONF libopenjp2)
 
 # Look for the header file


jriesmeier:
Thank you for the patch. Wouldn't it make more sense to replace "REQUIRED" with "QUIET"? Otherwise the entire configure process will fail if the CMake module "PkgConfig" is missing on the system.

CC: @eichelberg

jriesmeier:
Fixed with commit 82fb238.

agravgaard:
Sure, that's fine with me. But "Unknown CMake command" is also an error, so I think it would halt anyway in that case.

Thanks!

jriesmeier:
Right, therefore my committed patch looks like this:

```
diff --git a/CMake/FindOpenJPEG.cmake b/CMake/FindOpenJPEG.cmake
index 69617c6..8589edc 100644
--- a/CMake/FindOpenJPEG.cmake
+++ b/CMake/FindOpenJPEG.cmake
@@ -7,7 +7,10 @@
 #  OPENJPEG_FOUND        - True if OpenJPEG found
 
 # Use pkg_check_modules to determine the paths for OpenJPEG
-pkg_check_modules(OPENJPEG_PKGCONF libopenjp2)
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_check_modules(OPENJPEG_PKGCONF libopenjp2)
+endif()
 
 # Look for the header file
 find_path(OPENJPEG_INCLUDE_DIR NAMES openjpeg.h HINTS ${OPENJPEG_PKGCONF_INCLUDE_DIRS})

```

agravgaard:
Great, much better!

jriesmeier:
The new version will be visible on GitHub soon (still have to wait for the nightly builds).

PR #10, opened 2018-12-08
dcmdata/libsrc/dcrleccd: Minor clean up

FreddyFunk:
Declare the for loop control variables only for the loop scope.


eichelberg:
Thank you for the contribution, which has been committed today as commit 2c1f6e300 and will, as usual, appear in the public git in a couple of days when it has passed our nightly builds.

PR #11, opened 2018-12-08
dcmdata/libsrc/dcrleccd: Fix multiplication result may overflow 'int' before it is converted to 'size_t'

FreddyFunk:
Multiplication result may overflow `Uint16` before it is converted to `size_t` or `Uint32`.

### DIFF
--- a/dcmdata/libsrc/dcrleccd.cc
+++ b/dcmdata/libsrc/dcrleccd.cc
@@ -118,13 +118,14 @@ OFCondition DcmRLECodecDecoder::decode(
     {
       DcmPixelItem *pixItem = NULL;
       Uint8 * rleData = NULL;
-      const size_t bytesPerStripe = imageColumns * imageRows;
+      const size_t bytesPerStripe = static_cast<size_t>(imageColumns) * static_cast<size_t>(imageRows);
 
       DcmRLEDecoder rledecoder(bytesPerStripe);
       if (rledecoder.fail()) result = EC_MemoryExhausted;  // RLE decoder failed to initialize
       else
       {
-        size_t frameSize = imageBytesAllocated * imageRows * imageColumns * imageSamplesPerPixel;
+        const size_t frameSize = static_cast<size_t>(imageBytesAllocated) * static_cast<size_t>(imageRows) 
+								* static_cast<size_t>(imageColumns) * static_cast<size_t>(imageSamplesPerPixel);
         size_t totalSize = frameSize * imageFrames;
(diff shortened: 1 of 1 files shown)


dcmdata/libsrc/dcrleccd.cc
```diff
@@ -465,11 +466,12 @@ OFCondition DcmRLECodecDecoder::decodeFrame(
 
     DcmPixelItem *pixItem = NULL;
     Uint8 * rleData = NULL;
-    const size_t bytesPerStripe = imageColumns * imageRows;
+    const size_t bytesPerStripe = static_cast<Uint32>(imageColumns) * static_cast<Uint32>(imageRows);
```
hjmjohnson: I think you should use static_cast<size_t> instead of staic_cast<Uint32> here.

bytesPerStripe result is size_t.

dcmdata/libsrc/dcrleccd.cc
```diff
@@ -465,11 +466,12 @@ OFCondition DcmRLECodecDecoder::decodeFrame(
...
     Uint32 numberOfStripes = 0;
     Uint32 fragmentLength = 0;
     Uint32 i;
-    Uint32 frameSize = imageBytesAllocated * imageRows * imageColumns * imageSamplesPerPixel;
+    Uint32 frameSize = static_cast<Uint32>(imageBytesAllocated`) * static_cast<Uint32>(imageRows)
```
hjmjohnson: I think that frameSize either needs to be size_t  and the static_cast changed to size_t OR  this code needs to be reverted.

hjmjohnson: NOTE:  (imageBytesAllocated`) <-- trailing apostrophe needs to be removed.

eichelberg:
A few comments here:
- In DCMTK, we avoid the direct use of static_cast<>. We use the OFstatic_cast() macro instead, which either becomes a static_cast (on modern compilers) or a C-style cast (on ancient compilers). Perhaps not necessary anymore, but for reasons of coding style consistency we prefer this.
- Uint32 is safe for frameSize because the size of a single frame in DICOM is by definition smaller than 2^32 because no attribute value can be larger than 2^32-2 bytes, and all uncompressed pixel data is one attribute.
- I am not 100% sure if the explicit casts are really needed here or if C++'s integral promotion and integral conversion rules would be sufficient, but in any case they cannot hurt.
I have committed this patch (with amendments as described above) now as commit 29c610ead. This will, as usual, appear in the public git in a couple of days when it has passed our nightly builds.

PR #12, opened 2018-12-20
Fix for CMake exports and unified find mechanism

mschmieder:
This pull request addresses some issues that come from the CMake usage of this project. 

### Full paths in cmake exports
Currently DCMTK does not make use of CMake's import targets and therefore creates unusable cmake export files that contain full paths to the linked libraries. This is considered harmful and error prone since consuming DCMTK on another system than the one it was build will result in errors.

### Unified find mechanism
It is stated within the 3rdparty.cmake file that usually DCMTK is used with pre-compiled dependencies, the system should still allow for using the same mechanism on all hosts, including Windows.
I created a new option DCMTK_UNIFIED_FIND that, when enabled, goes through the same find mechanism than Unix/Apple would.

All changes should be backward-compatible.


mschmieder:
@hjmjohnson 
FindIconv.cmake was removed due to several issues.
  * FindIconv.cmake does not follow the proper naming convention of ICONV (all capital letters)
  * This leads to unexpected behavior on systems with non-case-sentitive file systems like Windows/MacOS where CMake already provides a **FindICONV.cmake** function out of the box. The fact that you are **appending** your CMake module path to **CMAKE_MODULE_PATH** makes the system select CMake's version primarily.

I'll try my best to resolve the current merge conflict and add more into the commit message.

It would also be great if DCMTK would provide some proper CI (Travis?) integration so changes of this kind can be checked against your references. Please consider integrating something like this in the future.

Cheers

hjmjohnson:
@mschmieder Thanks! To be clear,  I am not part of the DCMTK dev team, but I am very interested in the success of this project.   Your change is an important improvement for others who want to incorporate DCMTK into their toolkits easily.

offis-jas:
> FindIconv.cmake was removed due to several issues.
> 
>   * FindIconv.cmake does not follow the proper naming convention of ICONV (all capital letters)

Well, the proper naming convention of iconv seems to be all lower case letters,
writing that in camel case leads to FindIconv.cmake.

>   * This leads to unexpected behavior on systems with non-case-sentitive file systems like Windows/MacOS where CMake already provides a *FindICONV.cmake* function out of the box. The fact that you are *appending* your CMake module path to *CMAKE_MODULE_PATH* makes the system select CMake's version primarily.

Which is what we inteded by doing so. CMake's FindICONV.cmake is better than
ours, ours is only used as a fallback in case an old CMake version that does
not have it is used.
You should perhaps look at https://github.com/DCMTK/dcmtk/commit/0178bca95ab2b7ed89e2cb8d491da8b6649fbd3a for further details.

> It would also be great if DCMTK would provide some proper CI (Travis?) integration so changes of this kind can be checked against your references. Please consider integrating something like this in the future.

We are currently using a custom CDash setup for this purpose, but are
considering to integrate additional CIs as well.

PR #15, opened 2019-02-14
Header cleanup

eborisch:
Move to using `DCMTK_PACKAGE_VERSION`, `DCMTK_PACKAGE_STRING`, etc. rather than the generic `PACKAGE_VERSION` and friends. This is in keeping with a number of (most?) other packages that expose their versioning with #define-s that include their name.

Removed the autotools-based tooling in the process. I would suggest picking one build system and committing to it to avoid corner cases of "oh, yeah, we forgot to update the other build tool." (Plus then I didn't have to debug if these changes bothered the auto* tools at all.)

The resulting changes are evident in the installed files config/osconfig.h and dcmdata/dcuid.h. Builds & tests succeed in my testing.


jriesmeier:
Thank you for the pull request (patch). Unfortunately, you did not consider the Autoconf system that is still actively used by the DCMTK (in fact, you've removed the Autoconf support completely), so we cannot accept your patch in its current state.

eborisch:
I did consider it; and I decided to remove it ;) ...  If I have time someday I'll revisit this and look at what impact/changes are required to the autoconf flavor.

PR #17, opened 2019-06-10
Support for optional fractional part of a second

PavelTrushkin:
i have reworked OFTime::setISOFormattedTime and modified OFDateTime::setISOFormattedDateTime

new parser is very simple - old school char by char state machine :) but it works

Thanks!


jriesmeier:
Thank you for your contribution! Not sure whether we will apply the patch directly (or adjust it a little bit...) but in any case you'll be mentioned as the original author/contributor.

Since we currently have no time to check your patch in detail, I've added the following entry to DCMTK's issue tracker (so your proposed patch does get lost): https://support.dcmtk.org/redmine/issues/895

PavelTrushkin:
Thats okay let me know if i can help with the patch adaptation

PR #19, opened 2019-11-04
Fix TOC-TOU race condition for "-su --fork"

gsmethells:
Fix a race condition that occurs when both `-su` and `--fork` are used together. Multiple child processes sent the same Study Instance UID cause aborted Associations unnecessarily. See also https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use


gsmethells:
This is in the development phase for our orthopedic clinics

gsmethells:
@jriesmeier this has been tested and works to fix the race condition when both `-su` and `--fork` are used and there is more than one Association sending DICOM files with the same Study Instance UID. All other scenarios were left as-is so as to not change the behavior of other situations.

Please consider this change for inclusion with the upstream dcmtk code base. Let me know. Thanks.

Greg Smethells
Medstrat CTO
[www.medstrat.com](https://www.medstrat.com)

gsmethells:
@offis-jas any comment?

gsmethells:
@jriesmeier this is discouraging

PR #21, opened 2020-03-03
CMake normally uses /MT (not /MTd) in RelWithDebInfo configuration

dzenanz:


eichelberg:
Thank you for the pull request. The code you have modified in your patch does not exist anymore; it was removed with commit 82a221881 on 2020-02-25. The commit description is:

 Improved CMake options for the WIN32 build model.

When compiling on windows, two CMake options can now be used to control the Win32 build model (multithreaded or multithreaded DLL). When DCMTK_OVERWRITE_WIN32_COMPILER_FLAGS is ON, the build model can be selected by setting DCMTK_COMPILE_WIN32_MULTITHREADED_DLL (ON for multithreaded DLL, OFF for multithreaded). When DCMTK_OVERWRITE_WIN32_COMPILER_FLAGS is OFF, the DCMTK_COMPILE_WIN32_MULTITHREADED_DLL option is ignored and CMake's default compiler options are used.

We don't manually set compiler flags for Visual C++ anymore, so your issue should also have disappeared now.

PR #24, opened 2020-08-04
Make the TIFF metainfo closer to the PNG metainfo

malaterre:


jriesmeier:
Thank you for the proposed enhancement. The required changes have been committed. See commit ffd9307.

PR #25, opened 2020-08-13
Clarify indent of --grayscale option

malaterre:


jriesmeier:
Thank you for the hint. Improved documentation with commit f86acc7.

PR #27, opened 2021-02-15
Fix reading partial objects

akaraivanov:
This PR fixes reading DcmFileFormat when object is only partially available in input stream.


jriesmeier:
Thank you. We've added this issue to out to-do list so it doesn't get lost.

eichelberg:
The first issue, "Fix bug in DcmMetaInfo::read() when reading incomplete in stream object", was fixed in DCMTK commit #3e996a274 in March 2022, due to an another bug report including the same issue.

eichelberg:
The second issue, "Fix bug in DcmMetaInfo::read() when reading incomplete in stream object", was fixed today in DCMTK commit #5d8326524. This should appear in the public git repository in a couple of days.

PR #28, opened 2021-02-17
Fix TOC-TOU race condition for "-su --fork"

gsmethells:
Fix TOC-TOU race condition created when "-su --fork" are used together.

### DIFF
--- a/dcmnet/apps/storescp.cc
+++ b/dcmnet/apps/storescp.cc
@@ -32,6 +32,9 @@ BEGIN_EXTERN_C
 #ifdef HAVE_SYS_STAT_H
 #include <sys/stat.h>
 #endif
+#ifdef HAVE_ERRNO_H
+#include <errno.h>
+#endif
 #ifdef HAVE_FCNTL_H
 #include <fcntl.h>       /* needed on Solaris for O_RDONLY */
 #endif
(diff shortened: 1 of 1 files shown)


gsmethells:
@jriesmeier over a thousand orthopedic clinics and surgeons are awaiting your response to this.

gsmethells:
This is in live clinical use at a significant portion of the United States' orthopedic clinics. This has been field tested for several years now. I believe it is time to accept it into the mainline branch.

eichelberg:
I can understand that you are frustrated to still see this pull request open after two years. We are a small team, and as much as I'd love to see pull requests (and our issue tracker at https://support.dcmtk.org/redmine/issues/) processed in a timely manner, we currently simply don't manage that. The good news is that I have been assigned to your pull request now. Is this pull request a complete port of your prior pull request #19 (which consisted of 6 commits) to DCMTK 3.6.6, or is part of the old pull request still relevant?

gsmethells:
@eichelberg good to hear from you. I can understand your small team situation because we are one, too. The 1 commit here that has been squashed and re-based on version 3.6.6 is equivalent to pull request #19 as you already discerned. 

We recently moved up a version to 3.6.6 so I needed to re-base my changes and decided to attempt to garner some attention. Sorry to be yelling so loud.

gsmethells:
Full disclosure: this has been running for sometime now on several versions of CentOS Linux but has never been tested on Windows.

eichelberg:
Thanks for the warning re Windows. Unfortunately the storescp code has grown into a complex mess over the years. Originally, the tool was called "scp" and was intended as the simplest DICOM SCP implementation possible with DCMTK. Well, we're far beyond that point ;-)
So there is a real risk that fixing one thing might break another.

eichelberg:
I have had a look at the code now. The TOC-TOU race condition is indeed obvious if you run storescp with "-su --fork".
I have one question, though. Your patch accepts an existing subdirectory only if opt_sortStudyMode ==  case ESM_StudyInstanceUID. I agree that a race condition is rather unlikely in the other sort modes since the directory names will contain a timestamp with millisecond precision in that case. Nevertheless, unlikely as it may be, it is still possible, isn't it? Should the code not accept an existing directory in all sort modes (except for ESM_None of course, in which case this code is not executed anyway)?

gsmethells:
I agree that I was probably focusing the solution on our use case and that a race condition is likely for more than just that case. You know the code better than I do so I am open to changing the solution to another more informed one. I will set a reminder to look at this on Monday.

eichelberg:
I have merged your pull request (with minor modifications for Windows) into DCMTK today as commit #8036ce889. This should appear in a couple of days in the public git repository.

gsmethells:
Sounds great. Thanks @eichelberg

PR #29, opened 2021-02-25
Clarify JPEG vs JPEG-LS

malaterre:


jriesmeier:
Thank you. Fixed with commit e607a0598, which will be visible in the public git repository soon.

PR #30, opened 2021-03-24
fix compile error on Apple M1

Staars:
For the moment I see no other option than to switch some lines off for the M1.

Maybe SSE2NEON can supply the missing functions in the future.


michaelonken:
I've integrated the patch to DCMTK's internal testing branch (commit 5fba85). It will be visible on DCMTK master within the next days. Thank you for the report and patch.

PR #31, opened 2021-04-05
fix: cmake fix CMP0115 explicit file extensions

kayarre:
Currently because the cmake doesn't include the file extensions and therefore generates a lot of warnings with the latest cmake version 3.20.0

This modifes the cmake to be explicit.

```
Call Stack (most recent call first):
  dcmdata/libsrc/CMakeLists.txt:5 (DCMTK_ADD_LIBRARY)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMake/dcmtkMacros.cmake:82 (add_library):
  Policy CMP0115 is not set: Source file extensions must be explicit.  Run
  "cmake --help-policy CMP0115" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning
```


eichelberg:
As a stop-gap measure, we have set policy 115 to OLD with commit #736c7ae45 (on 6 April), but I agree that your patch is the better long-term solution. Will merge as time permits.

eichelberg:
I have now committed your pull request to the testing branch as commit #d090b6d7c. If everything works well, the commit should be visible in the public git in a couple of days.

PR #32, opened 2021-05-25
chore: auto download support libs for windows

maidamai0:
Auto download support libs for windows
* Add a python script that can be used to download support libs from [support website](https://dicom.offis.de/download/dcmtk/)
* This script can be integrated to `CMake` to auto-download at configuring time


jriesmeier:
Thank you for the suggested patch. I guess we would prefer to use CMake for this purpose instead of Python (in order to avoid another dependency).

PR #33, opened 2021-06-25
Too strong interpretation of PS3.3 / C.7.6.1.1.5 Lossy Image Compression

malaterre:
While browsing the DCMTK code base, I stumble upon the following lines:

http://git.dcmtk.org/?p=dcmtk.git;a=blob;f=dcmjpls/libsrc/djcodecd.cc;h=fc0df5c7f2bf60730e80638aef3491fcbc292804;hb=HEAD#l217

217     // set Lossy Image Compression to "01" (see DICOM part 3, C.7.6.1.1.5)
218     if (result.good() && (supportedTransferSyntax() ==
EXS_JPEGLSLossy)) result =
ditem->putAndInsertString(DCM_LossyImageCompression, "01");

This implementation is too strict per interpretation of PS 3.5 Section
A.4.3 JPEG-LS Image Compression.

The note #2 explicitly states that Transfer Syntax for Near Lossless
may be used:

[...]
Note that this process can, at the discretion of the encoder, be used
to compress images with an error constrained to a value of zero,
resulting in no loss of information.
[...]

In this case, a file decompressed by DCMTK even if originally lossless
(but using the transfer syntax "1.2.840.10008.1.2.4.81 "), will appear
to the rest of the world as lossy compressed:

Implication is that:

[...]
 It provides a means to record that the Image has been compressed (at
a point in its lifetime) with a lossy algorithm and changes have been
introduced into the pixel data. Once the value has been set to "01",
it shall not be reset.
[...]

I would suggest changing the signature of decodeFrame to also return
the max deviation as found in the JPEG-LS stream. Based on this value
(ie. != 0), then one can with confidence set the Attribute Lossy Image
Compression (0028,2110)

### DIFF
--- a/dcmdata/include/dcmtk/dcmdata/dccodec.h
+++ b/dcmdata/include/dcmtk/dcmdata/dccodec.h
@@ -129,6 +129,9 @@ public:
    *  @param decompressedColorModel upon successful return, the color model
    *    of the decompressed image (which may be different from the one used
    *    in the compressed images) is returned in this parameter.
+   *  @param isFrameLossless upon successful return, the lossless attribute
+   *    of the decompressed image (which may be different from the one used
+   *    in the transfer syntax) is returned in this parameter.
    *  @return EC_Normal if successful, an error code otherwise.
    */
   virtual OFCondition decodeFrame(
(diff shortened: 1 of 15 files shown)


malaterre:
There is very little interest in the use case. Also the implementation for JPEG 2000 would be "virtually impossible".

Closing.

PR #35, opened 2021-08-24
Document the expected format for VR AT

malaterre:
The String family function expects VR AT buffer to be formatted as
"(1234,5678)", with explicit parenthesis and explicit comma. Since
this is different from the JSON formatting, make sure to document it in
the header file.


jriesmeier:
Thank you for your proposed patch. I don't understand why you refer to JSON formatting but the format expected for the various VRs should be documented at the level of the VR classes, e.g. [DcmAttributeTag::putString()](https://support.dcmtk.org/docs/classDcmAttributeTag.html#ab9a5dec65e3d1858261199de31e6ad6e).

Maybe, we should add a general note to the putAndInsertXXX() method referring to the underlying implementation in the respective VR classes.

malaterre:
@jriesmeier Thanks for the quick review ! After too much reading of part18, I consider "(1234,5678)" to be the native XML format for VR AT, while "12345678" to the be the native JSON format:

* http://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_F.2.3.html

I do not know how to integrate your suggestion, so I am going to close this PR for now.

jriesmeier:
@malaterre DCMTK'S putAndInsertXXX() methods have nothing to do with DICOM's native XML or JSON format.

I will take care of the enhancement of the API documentation. Thank you for pointing out this missing detail.

PR #36, opened 2021-08-24
Provide a new API for DcmItem to delete a private attribute

malaterre:
Extend DcmItem with findAndDeletePrivateElement to allow user to delete
a private tag directly (reservation number is computed internally).


eichelberg:
Hi Mathieu, since you explicitly asked for my feedback - I am also not happy with the PR as it is now:

- The documentation talks about "deleting a private element", but apparently this is about passing the tag of a private creator element and then having the method delete that private creator and all associated private attributes. That is absolutely not clear from the API documentation.
- The mapping from private creator element to related private elements is managed through getPrivateCreator(), which is not the right way to do things since the private creator strings are not guaranteed to be unique and since the relationship is actually defined through the element tag, so the code does the wrong thing in a rather inefficient way :-)
- The searchIntoSub parameters makes no sense. Private elements in items have their own private creator elements and are not linked to one in the main dataset or a higher level (containing) sequence item

In general, however, I think a method that enables a user to "get rid" of a group of private elements by passing the tag of the private creator, or possibly the group number and the private creator string, would be useful to have.

PR #37, opened 2021-08-26
Provide a value for Character Encoding for stdlibc (iconv)

malaterre:
On Debian system, dcmtk is compiled with:

DCMTK_ENABLE_CHARSET_CONVERSION:STRING=stdlibc (iconv)

Which in turn reveal that:

$ dcm2xml --version
[...]
Character encoding: system default (unknown)

Update implementation detail to retrieve the character encoding using
the nl_langinfo() POSIX function.


michaelonken:
Closed with commit 7d78cf6598af1aa6a630cfab42cddea1d7ec7a15.

Thanks for the patch Mathieu.

PR #38, opened 2021-09-30
Prefer UTF-8 over latin1

malaterre:
this will be consistant with some other header files in DCMTK already
using UTF-8


jriesmeier:
Thank you for your report and suggested patch.  I personally would prefer to use the ASCII variant instead, i.e. transliterate the German umlaut "ü" to "ue". Otherwise, we would need character set conversion support for the coded entries or at least an indication that the code meaning is encoded in UTF-8.

By the way, which other DCMTK header files did you mean?

malaterre:
```
% find . -name "*.h" -exec file {} \; | grep -v ASCII
./dcmiod/include/dcmtk/dcmiod/modsopcommon.h: C++ source, UTF-8 Unicode text
./dcmseg/include/dcmtk/dcmseg/segment.h: C++ source, UTF-8 Unicode text
./dcmtls/include/dcmtk/dcmtls/tlsciphr.h: C++ source, UTF-8 Unicode text
./dcmdata/include/dcmtk/dcmdata/dcuid.h: C source, UTF-8 Unicode text
./dcmsr/include/dcmtk/dcmsr/codes/dcm.h: C source, UTF-8 Unicode text
```

jriesmeier:
Thanks again. I've just fixed the affected files and pushed the changes to DCMTK's internal testing branch.

PR #39, opened 2021-10-08
Reduce a warning level message to a single info level

malaterre:
Since commit 4ab084d3f the new way to disable external dictionary is to
use DCMTK_USE_DCMDICTPATH=OFF. However setting this configuration
triggers a warning message which may confuse the user. Instead reduce
the level to simply an info level. This variable is for advanced user
anyway. This makes the behavior consistant with the previous one.

Add new warning level messages for user transitioning from the previous
settings to the new settings.


michaelonken:
I pushed this (see comment in patch) to DCMTK testing, will be in master soon.

PR #41, opened 2021-10-12
Simplify code and remove explicit cast not required.

malaterre:
This removes gcc warning:

 error: useless cast to type ‘int’ [-Werror=useless-cast]

### DIFF
--- a/dcmimgle/include/dcmtk/dcmimgle/diovpln.h
+++ b/dcmimgle/include/dcmtk/dcmimgle/diovpln.h
@@ -553,11 +553,11 @@ inline int DiOverlayPlane::getNextBit()
 {
     int result;
     if (BitsAllocated == 16)                                        // optimization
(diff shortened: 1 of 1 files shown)


jriesmeier:
@malaterre Thank you for the report and proposed patch. However, I don't understand what is the purpose of fixing a single useless cast. When I enable -Wuseless-cast, I get hundreds of similar warnings. So, why do you want this particular cast to be fixed?

malaterre:
@jriesmeier I did not check other part of the toolkit for this particular warning. I simply fixed a single warning triggered by dcmtk codebase leaking into my application (inline function in header file). Maybe there are other way to hide this warning for user. I thought this was the right fix, since the other inline function did not explictly cast to int anyway.

jriesmeier:
@malaterre Thank you for the explanation. I've now fixed this issue in the current "testing" branch of the DCMTK. The changes should be visible in the public "master" in a couple of days.

jriesmeier:
Fixed with commit 495540c.

PR #42, opened 2021-11-03
Img2dcm jpegls

malaterre:


malaterre:
@michaelonken This is my current WIP. Do you have interest in this work ? If so, what is your preference for sharing duplicate code in between JPEG and JPEG-LS ? Thanks for your review.

michaelonken:
Hi Mathieu, thank you for sharing your patch with DCMTK upstream. I am not very familiar with JPEG-LS (I think Marco did the JPEG-LS part of DCMTK implementation) but scanning over your code, the JPEG-LS image source is very similar to the JPEG image source class? If most of the methods are re-used, maybe one should refactor it and put common functionality into a JPGUtil class?

malaterre:
I was tempted to create a `I2DJpegCommonSource` to share the duplicate code. I think it make sense to have two separate classes, one to handle JPEG and the other JPEG-LS. I'll try your suggestion to factor code in a new `JPGUtil` class. I'll update this PR directly.

michaelonken:
Yes, I would prefer the common functionality extracted into a `I2DJpegCommonSource` or `I2DJpegUtil` (which I would prefer but I leave this up to you). This class would then be used by the JPEG- and JPEG-LS-specific class. Such a refactored PR would be great. Thanks Mathieu.

malaterre:
@michaelonken I made another pass at this PR. Could you please do a quick review at least for the design. Before doing some further code cleanup, I'd like to get your feeling on the splitting of the logic.

If you take a closer look, you'll see that I've kept multiple commits, so that you see the time where I simply move code around (git is pretty smart at figuring this out).

Thanks for your time.

michaelonken:
I'll look into this, Mathieu

malaterre:
@michaelonken please do not. I need to seriously re-factor this (again!). My original design of splitting JPEG (itu81) from JPEG-Ls (itu-87) is non-sense.

michaelonken:
@malaterre Ok, let me know here once you're done

michaelonken:
Closed in favor of #50

PR #44, opened 2021-11-18
uses TRY_RUN check for signedness of char

malaterre:
dcmtk fails to cross build from source, because it uses a TRY_RUN check
to determine whether char is signed. Such a check does not have to run
code (which does not work during cross compilation), but can be written
as a compile only test. Please consider applying the attached patch to
fix this instance of TRY_RUN. There are more TRY_RUN to be fixed to make
dcmtk cross buildable though. So please consider applying the attached
patch as an incremental step.


malaterre:
Turns out this is now integrated upstream as:

* http://git.dcmtk.org/?p=dcmtk.git;a=commitdiff;h=99849fffe1771096a28e6ca6cae60017b8bfa72b

PR #45, opened 2021-11-22
Fix a compilation error in DSRListOfItems

malaterre:
This is triggered when using DCMTK_ENABLE_STL:BOOL=ON and a c++14
compiler.

  % cat t.cxx
  #include <dcmtk/dcmsr/dsrtlist.h>
  int main()
  {
    DSRListOfItems<int> l;
    l.removeItem(0);
  }

Gives:

  dsrtlist.h:303:30: error: cannot convert 'std::__cxx11::list<int, std::allocator<int> >::iterator' to 'std::__cxx11::list<int, std::allocator<int> >::const_iterator&'

### DIFF
--- a/dcmsr/include/dcmtk/dcmsr/dsrtlist.h
+++ b/dcmsr/include/dcmtk/dcmsr/dsrtlist.h
@@ -299,7 +299,7 @@ template<class T> class DSRListOfItems
     OFCondition removeItem(const size_t idx)
     {
         OFCondition result = EC_IllegalParameter;
-        OFLIST_TYPENAME OFListIterator(T) iterator = ItemList.begin();
+        OFLIST_TYPENAME OFListConstIterator(T) iterator = ItemList.begin();
         if (gotoItemPos(idx, iterator))
         {
             ItemList.erase(iterator);


jriesmeier:
@malaterre As far as I can see, there is no need to include the template class DSRListOfItems directly into a program. When DCMTK is compiled with STL enabled and CMAKE_CXX_STANDARD set to "14", I don't get any error like the above.

Of course, I can and also will change the type of the variable "iterator" in both removeItem() and insertItem() to "OFListConstIterator" if it causes trouble to you (for your use case). I was just wondering why you stumbled over it.

malaterre:
@jriesmeier The actual compilation issue was in dicomscope:

* https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1000220#5

My small toy example was just a reproducer.

jriesmeier:
Thank you for the clarification. I've just committed the required changes to the (internal) "testing" branch, which will be visible in the public "master" branch soon.

jriesmeier:
Fixed by commit 8f6c9ba0.

PR #47, opened 2021-12-14
Indicate that the input DICOM/JPEG-LS is not conforming

malaterre:


malaterre:
To reproduce the issue, I used samples dataset from:

* https://github.com/team-charls/charls/tree/master/test/jlsimage

Pick banny_HP1.jls, banny_HP2.jls or banny_HP3.jls

malaterre:
For reference:

* https://github.com/team-charls/charls/blob/master/include/charls/charls_jpegls_encoder.h#L84

> Color transformations are a HP extension and not defined by the JPEG-LS standard and can only be set for 3 component

while:

* https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_A.4.3.html

only reference ` [ISO/IEC 14495-1] `

eichelberg:
Good catch. I have merged this into the DCMTK testing branch now. Thank you!

PR #48, opened 2021-12-15
Document mechanism to add definition in CMake

malaterre:


malaterre:
To whoever read this PR, I think reference to `config/Makefile.def` may also need to be updated. Thanks (as usual) !

PR #52, opened 2022-03-04
Draft proposal for DcmScu::sendNCREATERequest

jogerh:
This is a draft to discuss options for implementing support for sending N-CREATE requests to an SCP. I would like to get feedback on this before potentially implementing a corresponding N-SET function.


michaelonken:
Hi Jøger, this is an excellent patch. I reviewed it and I only have a few minor remarks, if you're interested:

- One Question from a comment: Is it always right to assign abstract syntax to AffectedSOPClassUID, or are there scenarios where the client should provide an AffectedSOPClassUID that is different from the abstract syntax?
  - MO: It's always identical, so just copying it is fine.
- Question from a comment (edited): Is it ok to assign the status before checking the response?
  - MO: I think it is at least safe and we must return something. As long as the documentation states that return .bad() from the method makes all other return variables possibly invalid, it should be fine. I added a related note to the API documentation of the method.
- Made sure that dataset is received no matter whether caller wants it or not (in the latter case it is just deleted).
- Changed C++11 initialization (T_DIMSE_Message response = {};) to old style initialization (DCMTK does not presumes C++11 or later)
- Note: reqDataset is optional in standard, i.e. must not be present in every service. However, I don't know any DICOM service which uses N-CREATE without request dataset, so I keep it as is.

Attached the patch with the changes I made based on these comments. We are in the middle of doing a DCMTK release, so I am not sure that the patch will go into master before the release, that's why I did not push it upstream so far. But the patch will go into DCMTK before or after the release (scheduled for beginning of next month). So you could use it as is and later switch to the official master, if you are in a hurry right now.

Michael
[0001-Reviewed-patch.txt](https://github.com/DCMTK/dcmtk/files/8323856/0001-Reviewed-patch.txt)

jogerh:
Thank you for the review @michaelonken. Are the tests also considered for master?

LonnyChason:
What a brilliant patch, I'm trying to implement MPPS service using DCMTK, but found that DcmSCU has sendACTIONRequest, sendCANCELRequest, sendECHORequest, etc., but neither sendCREATERequest nor sendSETRequest. Really looking forward to this pull request being merged into the master.

michaelonken:
We'll finishing the release these days, won't take much longer :)

michaelonken:
Merged into DCMTK as c6fd8a, will shop up on master in a few days. Thank you for your contribution!

PR #53, opened 2022-03-14
Added check on no Unique Keys found in "strict mode".

heiwiper:
**Description**

When starting the dcmqrscp server using the -XF option and performing C-FIND query that has no keys in the levels above the Query/Retrieve Level, the returned error code is :

`0xc000: Failed: Unable to process`

This error code is caught after the function that strictly checks the query's conformity.

It can be tested by the following :

`findscu localhost 22222 -P -k QueryRetrieveLevel=STUDY -k StudyDate=20220314`

**Expected behavior**

The error code that should be returned according to the current implementation is the following :

`0xa900: Error: Data Set does not match SOP Class`


jriesmeier:
Thank you for your report and suggested fix. The changes have been tested and committed (see commit e3f899b0).

PR #54, opened 2022-03-18
Properly document default SOP Class for img2dcm

malaterre:


jriesmeier:
Thank you for the report. Fixed with commit 316822f08 (including another issue regarding "default options").

PR #56, opened 2022-03-21
Rework un-optimized huffman table checking of symbol table

malaterre:
Output a message that the huffman table contains a possibly un-optimized
symbol. This may hint the user for something like `jpegtran -optimize`.

This reverts partially commit deeda9adbcc148003aa36dc336fcefa4e004cbb2.


malaterre:
@eichelberg Hopefully I did get it right this time. Technically we could imagine one day to have a `dcmjpgtran` to provide functionalities similar to `jpegtran -optimize`. In which case it makes sense to hint user with a debug message.

Thanks for your comments

eichelberg:
I like this patch. One one hand it permits the decoding of images containing an unoptimized Huffman table, on the other hand it prints warnings to the debug logger when such a table occurs. I have merged this into our testing branch.

PR #58, opened 2022-03-24
Rework dcmdjpls log information to cope with CP-1843 changes

malaterre:
In bug #882, the behavior of dcmcjpls was changed to always set the
Planar Configuration to 0 (commit 55a8a0ff72). This commit update
behavior of dcmcjpls accordingly to that change.


jriesmeier:
Thank you for your proposed patch. I've slightly modified it, tested it and committed the changes to the "testing" branch.

The changes should be visible in the public git repository soon.

jriesmeier:
Fixed with commit 6913bda94.

PR #61, opened 2022-06-23
Make DcmSCU TCP connection timeout configurable per instance.

jogerh:
This change makes it possible to configure the DcmSCU TCP connection timeout per instance, instead of using the global dcmConnectionTimeout parameter.

In a multi-service executable, that for example supports QR, Worklist, and Storage, storage may be executed in parallel with Worklist and QR. This is a typical pattern on some DICOM modalities, where the user can do QR or Worklist queries while images are being sent in the background.

It is then beneficial to be able to configure different TCP connect timeouts for each individual service. A typical example could be a multi-site institution where the storage server is far away and requires a long TCP connection timeout, whereas the Worklist server could be on site and require a short TCP connection timeout.

This becomes particularly important for mobile devices, where the device may have access to a storage server, temporarily not have access to the worklist server. We would then like to have a short worklist timeout to quickly fall back to an offline worklist.


jriesmeier:
Thank you for your proposed patch, which looks pretty useful. However, when looking at the changes in detail, I wonder whether using "-1" for setting the timeout value from the global variable "dcmConnectionTimeout" really makes sense, because "-1" is already used for "unlimited".

jogerh:
Thank you for feedback @jriesmeier. You are absolutely right. Please have a look at the revised proposal that should address the ambiguity of the -1 value.

jriesmeier:
Thank you for the update. I just applied your proposed patches to my local development version of the DCMTK. There still seems to be a need for some adaptation in DCMTK module "dcmqrdb" and "dcmpstat". Did you disable these modules for your local build?

jogerh:
Yes, indeed. Thanks for testing this. I have now updated the remaining calls to ASC_createAssociationParameters from dcmqrdb and dcmpstat.

dcmnet/include/dcmtk/dcmnet/scu.h
```diff
@@ -1066,6 +1064,9 @@ class DCMTK_DCMNET_EXPORT DcmSCU
     /// ACSE timeout (default: 30 seconds)
     Uint32 m_acseTimeout;
 
+    /// TCP Connection timeout (default: global dcmConnectionTimeout
+    Uint32 m_tcpConnectTimeout;
```
jriesmeier, line 1068: @jogerh I guess this member variable should be of type Sint32, because a timeout of "unlimited" is specified by value of -1.

jogerh:
Absolutely right.

dcmnet/include/dcmtk/dcmnet/scu.h
jogerh: Absolutely right. Fixed in [2349204](https://github.com/DCMTK/dcmtk/pull/61/commits/2349204a3b7327de0a19c39c274468cdb3035746)

jriesmeier:
Another comment from my side (while further reviewing your proposed patch): DcmSCU::getConnectionTimeout() also needs to be modified since this method should now return "m_tcpConnectTimeout".

jogerh:
> Another comment from my side (while further reviewing your proposed patch): DcmSCU::getConnectionTimeout() also needs to be modified since this method should now return "m_tcpConnectTimeout".

Thank you for finding this issue @jriesmeier!

jriesmeier:
@jogerh I've now committed the first part of the proposed changes to our internal "testing" branch (with a few minor fixes). What is still missing are the new regressions tests, which I still hesitate to commit without further testing, because @michaelonken had so many problems with previous networking test cases on our nightly build servers.

jogerh:
That's great news! As for the tests, you could consider adding the tests that do not use a server but I will leave this up to you.

jriesmeier:
I will check the other parts of your PR later this week as I am currently pretty busy with other projects.

jogerh:
> @jogerh I've now committed the first part of the proposed changes to our internal "testing" branch (with a few minor fixes). What is still missing are the new regressions tests, which I still hesitate to commit without further testing, because @michaelonken had so many problems with previous networking test cases on our nightly build servers.

FYI https://github.com/DCMTK/dcmtk/pull/63 attempts to make DcmSCP better suited for running multiple unit tests in parallel on a build server.

jogerh:
Thank you for integrating this so quickly @jriesmeier.

jriesmeier:
@jogerh You are welcome. I will also have a look at the proposed regression tests (as time permits).

jriesmeier:
@jogerh wrote: "As for the tests, you could consider adding the tests that do not use a server..."
I just did exactly this, i.e. added two new test cases to "dcmnet/tests". Thanks again!

PR #63, opened 2022-06-28
Fix issue with port conflicts when running automated DICOM networking tests in a CI infrastructure

jogerh:
When running DICOM networking tests in a CI infrastructure, port conflicts can be a serious obstacle because only one test can run at a time. This prevents efficient utilization of CI hardware.

This change attempts to make DcmSCP more CI friendly by enabling automatic port assignment. This functionality is enabled by configuring the DcmSCP to listen to port 0. The underlying socket library will then assign an unused port to the DcmSCP when opening a listen port. The main part of this change is about propagating this automatically assigned port up to the API level.

This should prevent port conflicts when running multiple tests in parallel on the same OS instance.


michaelonken:
Integrated with commit d86bea (visible on master within the next days). Thanks for the patch!

PR #64, opened 2022-07-07
Korean (multi-byte) support ISO_IR 149

gridaine:
The goal is to support korean. 

Many Thanks to Sylvain C. for his help ;-)


jriesmeier:
@gridaine Actually, Korean (ISO_IR 149) is already supported by the DCMTK, but the [DICOM standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.12.html#table_C.12-4) requires to use it with "Code Extensions" according to ISO 2022, i.e. the Defined Term for Specific Character Set is "ISO 2022 IR 149".

jriesmeier:
A value of "ISO_IR 149" is not defined in the DICOM standard, so it makes no sense to support it. If Orthanc creates DICOM datasets with this value (for Specific Character Set), you should report this as a bug to the Orthanc team.

Also see this discussion: https://stackoverflow.com/questions/57173882/how-to-fix-lookuperror-unknown-encoding-iso-ir-149/57175833#57175833

PR #65, opened 2022-08-01
Fix typo in darwin CFLAGS assignment

mrbean-bremen:
- used CXXFLAGS instead of CFLAGS as base


jriesmeier:
Thank you. This issue (and a few others) are fixed now with commit 56a50fb48.

PR #66, opened 2022-08-01
Do not try to switch off warning if it does not exist

mrbean-bremen:
- warning 4351 only exists in VS versions before 2015


jriesmeier:
Thank you for your report and proposed patch. However, is it on purpose that you replaced "4521" by "4351"? The latter is already disabled a few lines later.

mrbean-bremen:
Ah, sorry - replaced the wrong line! Shall be corrected now.

jriesmeier:
@mrbean-bremen Thank you for your confirmation. I just committed the required changes to the "testing" branch of the DCMTK. If the nightly builds run without any errors (or warnings), the commit will be visible in DCMTK's public git repository soon.

PR #68, opened 2022-09-02
Allow changing ImplementationClassUID and ImplementationVersionName by providing the appropriate global variables.

doskachok:
This change make it possible for client's code to configure the values of  ImplementationClassUID and ImplementationVersionName. By default OFFIS values will be used.

It is needed to identify and distinguish implementation environments from each others, especially in case when the module is used by several product types.


jriesmeier:
Thank you for your contribution. I personally don't like the introduction of two new OFGlobal instances for this purpose. Why don't you extend the underlying ASC_createAssociationParameters() function accordingly, i.e. by two _optional_ parameters for the Implementation Class UID and the Implementation Version Name? This seems to be more natural to me as it is the function where the two values are copied to the T_ASC_Parameters structure.

malaterre:
> I personally don't like the introduction of two new OFGlobal instances for this purpose.

+1 You never understand why a call in a random function from your process affect some other random function from the same running process...

doskachok:
Thanks for the review! I will have a look how to do it with extending ASC_createAssociationParameters. At this moment I can say that it will require to add setters for these values at least in DcmSCU and DcmSCP. They are also used when writing files, not only in association, thus I will probably need to add setters there as well.

jriesmeier:
Right, the information is also used in the file meta information header (of the DICOM file format). Alternatively to setting the values at runtime, there could be a define that declares two extern const char * global variables/constants that have to be defined in the user code. This would make sure that the user program uses the same identifying information, but it would still allow to specify different information for different user programs bass on the same DCMTK library.

doskachok:
I just created draft PR that implements "optional parameters" suggestion, please have a look: #69

PR #69, opened 2022-09-07
Draft: Add possibility to change Implementation Class UID and Implementation Version Name

doskachok:
This PR implements suggestion from #68, which is the initial version.

This change make it possible for client's code to configure the values of ImplementationClassUID and ImplementationVersionName. By default OFFIS values will be used.

It is needed to identify and distinguish implementation environments from each others, especially in case when the module is used by several product types.


eichelberg:
Fixed a few minor issues and merged the pull request into DCMTK's testing branch.

PR #70, opened 2022-10-04
Make DcmSCU::nextMessageID protected instead of private

jogerh:
DcmSCU provides several virtual functions where the intent is to enable users to override these virtual functions. For example, DcmSCU::sendMOVERequest advertises that it can be overridden in the doc string.

To be able to override these functions in a meaningful way, we need access to the nextMessageID function from derived classes.

A use case for overriding DcmSCU::sendMOVERequest could be to enable concurrent C-MOVE requests. In such cases, we need a way to get the message ID out of the DcmSCU to be able to correlate it with the Move Originator Message ID from a client side STORE SCP.


michaelonken:
Closed with 81c0edbcb8da8b2e4d15f912bab6f07ad90c3bfc

PR #71, opened 2022-10-07
Move code related for network initialization in DcmBaseSCPPool::listen to the separate function and make it virtual.

doskachok:
This way it is possible for the clients to implement a notification about the results of the network initialization part. It is needed to be able to get feedback for the cases where,  for example, the selected port is already taken.


michaelonken:
Integrated this into testing branch, will be available on master within the next days.

michaelonken:
Closed with cdb7182633af03ec78691c17b1c0e707b5c4ef8e

PR #74, opened 2022-12-12
Prevent macro-redefined warning for '__STDC_FORMAT_MACROS'

chrisse74:
When compiling in tandem with gRPC we experienced the warning: '__STDC_FORMAT_MACROS' macro redefined [-Wmacro-redefined]. On gRPC side the define is checked prior to setting it, this seems to make sense here too.


eichelberg:
I just committed the change as DCMTK commit #5405b5d37.

PR #76, opened 2023-01-10
Make it possible to specify MessageIDBeingRespondedTo for sendCANCELRequest.

doskachok:
This is needed to be able to use sendCANCELRequest implementation from DcmSCU without overriding it when nextMessageId() is overridden.


michaelonken (commented):
Thanks for the patch. msgIDBeingRespondedTo must allow positive 16 bit values. Maybe use Sint32 and make sure before using the value that it fits into Uint16 (Message ID) before casting.

michaelonken:
I merged this into DCMTK's internal testing branch. The related commit df69ef685 will show up on public DCMTK master within the next days.

PR #77, opened 2023-01-24
Make JSON output valid even if "pixel data is encapsulated"

reunanen:
Problem: JSON output as a whole is not valid JSON when "pixel data is encapsulated" (whatever that means)

Solution: even in this case, write JSON Opener and Closer


reunanen:
Before (not exactly valid JSON):
image

After (probably not very useful as such, but at least valid JSON):
image

eichelberg:
Merged into DCMTK testing branch today as commit #e8e312828. This will appear in a couple of days in the public git repository once it has passed our nightly builds and test suite.

PR #78, opened 2023-04-14
Provide an option to override default cmake search logic

malaterre:


eichelberg:
Merged into testing branch on 2023-04-21 as commit #5512aae0e.

PR #81, opened 2023-06-12
Fix detection of unsigned char on MSVC

malaterre:
cl would fail to compile the code with error:

error C2131: expression did not evaluate to a constant
note: a non-constant (sub-)expression was encountered


michaelonken:
Closed with commit 6d32aea619590b6beb5e04bf343b9268b05c5cfe.

PR #87, opened 2023-09-07
Fix handling of malicious PDV size

doskachok:
It was possible to overflow the _length_ variable in DT_2_IndicatePData on line "length -= 4 + pdvLength;" when pdvLength is set by a server to a value greater then LONG_MAX + pduLength. By doing that the _length_ overflowed and became positive, even if it was initially less than 4 + pdvLength. Due to the overflow "if (pdvLength < 2)" was not working and in the next loop iteration EXTRACT_LONG_BIG was causing read access violation.

Added check "length < 4 + pdvLength" to solve this problem.
Added check ULONG_MAX - pdvLength < 4 to make sure pdvLength + 4 will not overflow unsigned long.

### DIFF
--- a/dcmnet/libsrc/dulfsm.cc
+++ b/dcmnet/libsrc/dulfsm.cc
@@ -1387,8 +1387,7 @@ DT_2_IndicatePData(PRIVATE_NETWORKKEY ** /*network*/,
     unsigned long
         pduLength,
         pdvLength,
-        pdvCount;
-    long
+        pdvCount,
         length;
     unsigned char
        *p;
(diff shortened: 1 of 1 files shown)


michaelonken:
Since this is security-related, we decided to integrate this into the upcoming release. See commit 1549d8c. Thanks for your report and patch!

doskachok:
@michaelonken, that's great news, thank you! Could you please also have a look at #79 when you have time after release?

michaelonken:
Sure, w already have an internal feature branch that expanded on the work of #79. No pull request gets lost ;)

PR #89, opened 2023-10-19
Typo in the interface

kwach:
declaration:
const Uint16 length
vs
definition:
const Uint16& length

This results in linking error when trying to use it.


michaelonken:
Thank you for the report, this will be fixed in commit aa4c5e available on GitHub in a few days, latest.

PR #90, opened 2023-10-20
Add userIdent negotiation to movescu app

kwach:
Just as it is done in storescu with the same params and all


michaelonken:
Thank you, I will check it in detail and let you know after the pending release.

kwach:
friendly reminder :D

PR #91, opened 2023-10-25
Fix CMake Config Generation for Cross Compiling

parasite-lost:
When cross compiling on GNU/Linux for target platform Windows use correct path separators during DCMTK's CMake config generation.

### DIFF
--- a/CMake/GenerateDCMTKConfigure.cmake
+++ b/CMake/GenerateDCMTKConfigure.cmake
@@ -142,7 +142,8 @@ else()
 endif()
 
 # Configure dictionary path and install prefix
-if(WIN32 AND NOT CYGWIN)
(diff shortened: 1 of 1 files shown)


mp-chet:
Tested in our MinGW cross-compilation pipeline on Linux; works fine.

eichelberg:
I have looked at the proposed change, and while I understand what you are aiming at, I am not sure if this is the right solution. The problem I see is that through this modification, some variables use forward slashes in path expressions that are later used within the DCMTK executables on Windows:

- DCMTK_DEFAULT_CONFIGURATION_DIR
- DCMTK_DEFAULT_SUPPORT_DATA_DIR
- DCM_DICT_DEFAULT_PATH (e.g. when compiling with ENABLE_PRIVATE_TAGS)

PR #93, opened 2024-01-31
Change Tiff and JPEG library vars to match CMake output

sameeul:
Changing `TIFF_LIBRARY` to `TIFF_LIBRARIES` and `JPEG_LIBRARY` to `JPEG_LIBRARIES`, which are the correct result variables of  `find_package(TIFF QUIET)` and `find_package(JPEG QUIET)` calls.

Source: 
https://cmake.org/cmake/help/latest/module/FindTIFF.html
https://cmake.org/cmake/help/latest/module/FindJPEG.html

I came across this as I was trying to build the libraries as a dependencies for another downstream package. My initial question was posted in the dcmtk forum (https://forum.dcmtk.org/viewtopic.php?t=5350&sid=ad563ed914d3ca3c19bd62c359ddc476)


michaelonken:
@eichelberg I did not check this in detail but at least the CMake's FindTiff.cmake module actually sets the TIFF_LIBRARIES variable (e.g. [in latest version](https://cmake.org/cmake/help/latest/module/FindTIFF.html)).  

Older versions state, that TIFF_LIBRARY is also set but not meant for general use. 

So overall I think TIFF_LIBRRARIES is the official way to go. I can merge this if you agree.

michaelonken:
I will integrate this soon.

michaelonken:
Closed by a92b79c7134eca0588d84817a6a1670e46016f4d.

PR #94, opened 2024-03-11
Support CMAKE_CROSSCOMPILING_EMULATOR

thewtex:
When set, CMAKE_CROSSCOMPILING_EMULATOR will be used to run try_run executables. DCMTK_TRY_RUN interferes with its operation. Also, do not warn about an emulator not being available when CMAKE_CROSSCOMPILING_EMULATOR is set.

Co-authored-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>

@michaelonken patch from ITK that improves cross-compilation, including for wasm.

CC @jcfr @fedorov


michaelonken:
Integrated with f3231c687. Will be showing up on public master within a few days (after passing internal dashboard builds).

PR #96, opened 2024-03-29
Document static global mechanism

malaterre:


jriesmeier:
Thank you for your PR. Adding an output like this to the debug logger seems to be a good idea. I will slightly modify the text of the message and then commit the changes to the "testing" branch.

jriesmeier:
Here is the corresponding commit b3fbdcc6162.

PR #97, opened 2024-04-18
Fix dereferencing zeroth element of empty vector

pbatko-ge:
This is undefined behavior. In MSVC results in failed assertion in debug build.

Note: `std::vector::data` is available since C++11. If you want to support pre-C++11 standards, this changes should be rewritten.


eichelberg:
A (simpler) fix that does not require C++11 was committed today as commit #169030f5a.

PR #99, opened 2024-06-04
IPV6  Unknown host error

SobhitaMercy:
Fixed IPV6 address association failures in requestAssociationTCP() function of dulfsm.cc. 
Reference https://forum.dcmtk.org/viewtopic.php?t=5373 

The issue was due to getaddrinfo() not checking IPV6 addresses for DNS lookup, resulting in an ‘Attempt to connect to unknown host’ error.  Additionally, corrected sscanf formatting issue for calledPresentationAddress when an IP6 address is passed.

### DIFF
--- a/dcmnet/libsrc/dulfsm.cc
+++ b/dcmnet/libsrc/dulfsm.cc
@@ -2255,8 +2255,22 @@ requestAssociationTCP(PRIVATE_NETWORKKEY ** network,
     int s;
 #endif
     struct linger sockarg;
-
-    if (sscanf(params->calledPresentationAddress, "%[^:]:%d", node, &port) != 2)
+	
+    // ipv6 address was not parsing properly, so corrected the below code.
+    // convert the char array to a std::string
+    std::string addressString(params->calledPresentationAddress);
+    // find the last colon in the string to split the node and port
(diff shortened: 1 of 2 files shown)


eichelberg:
I have addressed some portability issues in this PR and commited the result to DCMTK's testing branch today. Once it passes our nightly builds, the commit should become visible in the public Git repository in a few days. Thank you for your contribution.

eichelberg:
I have to reopen the pull request since the modification in OFStandard::getAddressByHostname() causes problems on some of our test platforms. Specifically, it can happen that a hostname such as "localhost" gets resolved to an IPv6 address. If the DICOM association acceptor on that host only supports IPv4 (as all DCMTK applications currently do), they cannot be reached anymore. Therefore, a more flexible mechanism that allows a client to select between IPv4, IPv6 or the automatic hostname resolution is required.

eichelberg:
I now have implemented a solution based on your PR that does not break backward compatibility. There is now a new function
ASC_setProtocolFamily() that allows the application to select the protocol family before an outgoing connection is requested. There are three possible values, corresponding to IPv4, IPv6 and automatic selection based on a DNS lookup (AF_UNSPEC). This will appear in a couple of days in our public git repository.

PR #100, opened 2024-06-12
Resolve ambiguous VR taken from dictionary

mrbean-bremen:
- with dcmPreferVRFromDataDictionary set, the VR is reset to the dictionary VR, which may be ambiguous (e.g. EVR_ox, EVR_px...); in this case, the VR has to be resolved

### DIFF
--- a/dcmdata/libsrc/dcitem.cc
+++ b/dcmdata/libsrc/dcitem.cc
@@ -1098,6 +1098,9 @@ OFCondition DcmItem::readTagAndLength(DcmInputStream &inStream,
         /* the VR in the dataset might be wrong, so the user can decide to ignore it */
         if (dcmPreferVRFromDataDictionary.get() && (newEVR != EVR_UNKNOWN) && (newEVR != EVR_UNKNOWN2B))
         {
+            /* resolve ambiguous VRs */
+            checkAndUpdateVR(*this, newTag);
+            newEVR = newTag.getEVR();
             if (newEVR != vr.getEVR())
             {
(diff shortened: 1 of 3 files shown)


mrbean-bremen:
As discussed in the forum - a call to `checkAndUpdateVR` as proposed by @jriesmeier fixed this.

jriesmeier:
@mrbean-bremen Thank you! This looks very good. I'll probably enhance DcmItem::checkAndUpdateVR() to return information on whether the VR has changed or not, but then your proposed patch should be ready for being committed.

I'll update this thread as soon as the commit has been pushed and is publicly available.

mrbean-bremen:
Thanks 😀 

An unrelated observation:
You have a GitHub actions workflow to run the tests (even if only under Windows), but that has to be approved to run. 
The default setting is "Require approval for first-time contributors", but given that any PR will be from a new contributor, as they are never merged here, the workflow never runs if not approved.

I'm aware that you have your own tests running in the original repo, but one of the reasons to make a PR here is to see the result of the tests. Maybe you can change the settings to "Require approval for first-time contributors who are new to GitHub", so the feature would actually be used?

Thanks!

jriesmeier:
Regarding your "unrelated observation": This is something @michaelonken should comment on, as he enabled this feature and is also more active on GitHub.

michaelonken:
👍 Sure, updated the Action permissions as proposed.

mrbean-bremen:
Thanks!

PR #101, opened 2024-07-05
Reduce log level in readUntilTag

malaterre:
No need to repeat at information level what the programmer has setup.


jriesmeier:
Thank you for your report and proposed patch. I've committed the required changes to the internal "testing" branch. Commit 2d0e86ede should be visible in DCMTK's public git repository rather soon (in 1 or 2 days).

PR #102, opened 2024-07-09
Fixed ofstd build error in debug mode with C++20.

khangthk:
Found an issue when building ofstd in debug mode with c++20.
image

### DIFF
--- a/ofstd/libsrc/ofcmdln.cc
+++ b/ofstd/libsrc/ofcmdln.cc
@@ -1243,7 +1243,7 @@ OFCommandLine::E_ParseStatus OFCommandLine::parseCommandFile(const wchar_t *argV
 #ifdef DEBUG
             if (block != 0)
             {
-                ofConsole.lockCerr() << "WARNING: closing quotation mark (" << block << ") missing in command file " << strValue << OFendl;
+                ofConsole.lockCerr() << "WARNING: closing quotation mark (" << (unsigned)block << ") missing in command file " << strValue << OFendl;
                 ofConsole.unlockCerr();
             }
 #endif


michaelonken:
Thank you for the report and suggested patch. Pushed upstream via commit fcc01596e (publicly available within the next days).

jriesmeier:
Personally, I don't like the proposed (and not committed) change. First, we don't use C-style typecasts anymore in the DCMTK. Second, converting a wchar_t to int output the numeric character code but not the character as for the char* variant of OFCommandLine::parseCommandFile(). The preferred solution would be: either convert wchar_t to char for the output message, or use an appropriate output stream that supports wchar_t.

michaelonken:
We can make an OFstatic_cast from it regarding your first point. Regarding the second: I think printing the character code for debugging purposes is acceptable.  I agree that if there is better patch that does a real conversion we can use that one instead.

jriesmeier:
To me, the current change, which was actually committed, looks like a "quick hack" that was done to keep the compiler silent. This is not how it should be done (see above). Just my 2 cents.

michaelonken:
Until we have something better, shall we use
`ofConsole.lockCerr() << "WARNING: closing quotation mark (wchar_t character code " << OFstatic_cast(unsigned, block) << ") missing in command file " << strValue << OFendl;`?

Since you are the original author of the file, we can also revert the change overall and create a ticket, fine for me either way.

jriesmeier:
I would suggest to change the current implementation to use the OFstatic_cast() typecast and add a comment to the source code that this is just an "interim solution" (or the like) and that the code should actually print the character and not the associated code. And, you should also create a ticket in our issue tracker. Thanks.

michaelonken:
For reference: Related [Issue in DCMTK tracker](https://support.dcmtk.org/redmine/issues/1130).

PR #104, opened 2024-07-13
Allow linking to extra libs static when not using find_package()

khangthk:
I build static libxml2 library with latest code, when building dcmtk with BUILD_SINGLE_SHARED_LIBRARY=ON it gives an error because 'bcrypt.lib' is not linked to. The library 'bcrypt.lib' is required by libxml2 ([see here](https://github.com/GNOME/libxml2/commit/7e83a089949184f8c0df3f9985bf0876fdecd5b6)).

After some searching, I found that LIBXML2_EXTRA_LIBS_STATIC can be used to add the 'bcrypt.lib' library.
But this item is not available when DCMTK_USE_FIND_PACKAGE is OFF

image


michaelonken:
Thank you for the report. Where is  LIBXML2_EXTRA_LIBS_STATIC defined?

I downloaded the latest code of libxml2 via https://github.com/GNOME/libxml2. It seems it does not define LIBXML2_EXTRA_LIBS_STATIC.

Also in your patch I noticed that you add EXTRA libs to other packages, like OPENSSL_EXTRA_LIBS_STATIC. Why?

khangthk:
> Thank you for the report. Where is LIBXML2_EXTRA_LIBS_STATIC defined?
> 
> I downloaded the latest code of libxml2 via https://github.com/GNOME/libxml2. It seems it does not define LIBXML2_EXTRA_LIBS_STATIC.
> 
> Also in your patch I noticed that you add EXTRA libs to other packages, like OPENSSL_EXTRA_LIBS_STATIC. Why?

XXX_EXTRA_LIBS_STATIC has been defined and used before in dcmtk. You can search in the source code with the keyword "_EXTRA_LIBS_STATIC"
When building dcmtk I will set this value, for example: cmake.exe -D LIBXML2_EXTRA_LIBS_STATIC=bcrypt.lib

michaelonken:
Oh, wow, you're right, for some reason I never realized those in DCMTK :smile: I will get back to you.

khangthk:
There is one more issue I have not mentioned.
In the file [dcmtk\CMake\3rdparty.cmake](https://github.com/DCMTK/dcmtk/blob/master/CMake/3rdparty.cmake) there is a difference regarding the use of the 'LIBXML_STATIC' macro when the DCMTK_USE_FIND_PACKAGE option is ON and OFF.
I summarize the differences in that file as follows:

```
if(DCMTK_USE_FIND_PACKAGE)
    if(DCMTK_WITH_XML)
        if(NOT LIBXML2_FOUND)
            ......
        else()
            .......
            set(WITH_LIBXML 1)
        endif()
    endif()
else()
    if(DCMTK_WITH_XML)
        if(WITH_LIBXMLINC)
            ......
            set(WITH_LIBXML 1)
            # this hides some warnings that are emitted when linking against libxmlXXX.lib instead of linking the DLL directly
            add_definitions("-DLIBXML_STATIC")
        else()
            ......
        endif()
    endif()
endif()
```

When DCMTK_USE_FIND_PACKAGE is set to ON, the build fails because the LIBXML_STATIC macro cannot be found.
If I add line add_definitions("-DLIBXML_STATIC") after set(WITH_LIBXML 1) it builds OK. I only tested on Windows.

This is a new problem, my patch is only for case DCMTK_USE_FIND_PACKAGE=OFF.

PR #105, opened 2024-07-18
Fix single shared library build error in Linux

khangthk:
I encountered an error while building a single shared library on Linux (Ubuntu 24.04 LTS, GCC 13.2.0, CMake 3.30.0)
Screenshot from 2024-07-19 00-47-57

### DIFF
--- a/dcmjpeg/libijg12/jaricom.c
+++ b/dcmjpeg/libijg12/jaricom.c
@@ -12,4 +12,4 @@
 #include "jinclude12.h"
 #include "jpeglib12.h"
 
-IJG_INT32 jaritab[1];	/* dummy table */
+extern IJG_INT32 jaritab[1];	/* dummy table */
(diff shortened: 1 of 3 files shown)


michaelonken:
Thank you, I could reproduce the problem and took over the solution. Maybe @eichelberg  can comment whether we could remove jaricom.c completely in the future since DCMTK never supported arithmetic JPEG coding anyway.

eichelberg:
I am not happy with the modification. The only purpose of this dummy variable is to prevent linker warnings. On some platforms, the linker will issue a warning if one translation unit does not define any symbols. A proper solution would be to either name the variable differently in the three different files, or to remove jaricom.c completely. This file is indeed only a placeholder for the patch for the IJG library that adds support for arithmetic coding. We never implemented that in DCMTK because that patch is mutually incompatible with the patch that adds support for lossless JPEG, which we do use. Therefore, this file can be removed.

michaelonken:
I will remove the file then.

PR #106, opened 2024-07-25
Fix missing 'errno' definition.

Naum-PDFTron:
### Description
Compiling on Linux Docker image with gcc4.8 reports missing 'errno'. Included 'cerrno'.

### Changelog entry
N/A

### DIFF
--- a/ofstd/libsrc/ofchrenc.cc
+++ b/ofstd/libsrc/ofchrenc.cc
@@ -18,7 +18,6 @@
  *  Purpose: Class for character encoding conversion (Source)
  *
  */
-#include <cerrno>
 
 #include "dcmtk/config/osconfig.h"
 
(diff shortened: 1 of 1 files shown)


michaelonken (changes requested):
Does including <cerrno>also work for you if you put it below the osconfig.h include, e.g. before the windows.h include #ifdef?

We want osconfig.h to be the first include in any DCMTK file.

Naum-PDFTron:
It doesn't matter where it is put. I thought system includes go first. Here what we are getting:
Build-#15

michaelonken (approved):

michaelonken:
I realized that your branch is still using an old DCMTK commit from March.

Meanwhile the related file has already been fixed regarding errno by starting like this:

```
#include "dcmtk/config/osconfig.h"

#include "dcmtk/ofstd/ofchrenc.h"
#include "dcmtk/ofstd/ofstd.h"
#include "dcmtk/ofstd/ofdiag.h"
#include "dcmtk/ofstd/ofconsol.h"
#include <cerrno>
```

Thanks for the report anyway.

PR #107, opened 2024-08-01
Fix negative array access on EXS_Unknown

percontation:
When a `DcmXfer` is initialized with `EXS_Unknown` (which does happen), a negative array access happens on `XferNames`.

The existing `XferNames[i].xfer == xfer` check prevents anything too wild from happening, unless a build is super unlucky about what appears before `XferNames` in global memory. But, better not to rely on that.


jriesmeier:
Thank you for the report. We will make sure that the issue will be fixed rather soon.

jriesmeier:
The required changes have been committed and merged to the master branch: see commit 84c57f816.

PR #108, opened 2024-08-16
ENH: Replace HTML_HYPERLINK_PREFIX_FOR_CGI with std::string argument

thewtex:
The dsr2html tool uses a hardcoded literal string as a hyperlink prefix for refrenced composite objects (images, waveforms, etc.) in rendering the Key Object Selection documents.

Replace this literal with a string argument that will be passed by a calling application / function.

@michaelonken here is a patch that would be helpful to upstream.

CC @jadh4v @fedorov

### DIFF
--- a/dcmsr/libsrc/dsrwavvl.cc
+++ b/dcmsr/libsrc/dsrwavvl.cc
@@ -195,10 +195,11 @@ OFCondition DSRWaveformReferenceValue::writeItem(DcmItem &dataset) const
 OFCondition DSRWaveformReferenceValue::renderHTML(STD_NAMESPACE ostream &docStream,
                                                   STD_NAMESPACE ostream &annexStream,
                                                   size_t &annexNumber,
+                                                  const std::string& urlPrefixToCompositeObjects,
                                                   const size_t flags) const
 {
     /* render reference */
-    docStream << "<a href=\"" << HTML_HYPERLINK_PREFIX_FOR_CGI;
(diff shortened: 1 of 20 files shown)


dcmsr/libsrc/dsrwavvl.cc
```diff
@@ -195,10 +195,11 @@ OFCondition DSRWaveformReferenceValue::writeItem(DcmItem &dataset) const
 OFCondition DSRWaveformReferenceValue::renderHTML(STD_NAMESPACE ostream &docStream,
                                                   STD_NAMESPACE ostream &annexStream,
                                                   size_t &annexNumber,
+                                                  const std::string& urlPrefixToCompositeObjects,
                                                   const size_t flags) const
 {
     /* render reference */
-    docStream << "<a href=\"" << HTML_HYPERLINK_PREFIX_FOR_CGI;
+    docStream << "<a href=\"" << urlPrefixToCompositeObjects;
```
jadh4v, line 202: Do we need to change these lines to this?
```c++
docStream << "<a href=\"" << urlPrefixToCompositeObjects.c_str();
```
Or the current one is fine?

jadh4v (approved):
Just one minor point, otherwise looks good.

jriesmeier:
Thank you for your contribution. I've already checked your proposed patch and applied it locally with a few modifications (e.g. regarding DCMTK conventions, missing API documentation, order of parameters). I will probably also add a new command line option to dsr2html that allows for specifying the URL prefix for hyperlinks to composite objects.

PR #110, opened 2024-10-14
Take into account PhotometricInterpretation

tuket:
I have a DICOM file that can't be loaded because the PhotometricInterpreation is being cosidered correctly when the Image class is created.
[3DSlice52.zip](https://github.com/user-attachments/files/17365290/3DSlice52.zip)


jriesmeier:
Thank you for your PR. However, it is not needed as the DCMTK behaves as expected, e.g. when using dcmj2pnm to process the DICOM image. Since the DICOM image is compressed with JPEG Lossless, you need to register support for the JPEG decoders to your program (see e.g. source code of dcmj2pnm).

tuket:
Ah, true. My bad. Cheers!

PR #111, opened 2024-12-11
Properly pass the compressed buffer to fixPaddingIfNecessary

malaterre:


eichelberg:
Oops... embarrassing copy/paste error. And the culprit is... me.
I have just committed the fix as commit #bb650ebfe.

PR #112, opened 2024-12-13
ofstub.cc: fix environ for Apple

barracuda156:
Use the correct declaration on macOS.

### DIFF
--- a/ofstd/libsrc/ofstub.cc
+++ b/ofstd/libsrc/ofstub.cc
@@ -141,6 +141,11 @@ static OFString getLastErrorString()
   return OFString();
 }
 
+#elif defined(__APPLE__)
+
+#include <crt_externs.h>
+#define environ (*_NSGetEnviron())
+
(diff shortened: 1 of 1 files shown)


michaelonken:
Thanks Sergey, 

did you try this out? I could find [old references](https://www.manpagez.com/man/7/environ/osx-10.5.php) that it is required to explicitly define environ this way so I wonder whether this is still required and works on "all" Apple systems.

Thanks!

barracuda156:
@michaelonken Yes, this way works across the board, which can be seen from a number of ports on buildbots; and CI pass for the latest macOS versions specifically for this one with the fix applied: https://github.com/macports/macports-ports/pull/27063

I am not sure of the exact threshold when macOS started handling a Linux declaration for environ, but it does fail at least on some.

michaelonken:
Merged into upstream with f4d877 (visible on public muster branch within the next days).

barracuda156:
Thank you!

PR #113, opened 2025-01-03
Make `DcmList` random access be O(1)

reunanen:
Problem: Reading a real-world DICOM image was a bit slow, and simple profiling suggested that some 50% of the time (*) was spent in `DcmList::seek_to` (which has a random-access-like interface, but internally _iterates_ over a doubly linked list)

Solution: Internally use `std::deque` that provides fast O(1) random access

(*): the other 50% was spent decoding jpeg tiles, which is legit use of time of course

### DIFF
--- a/dcmdata/include/dcmtk/dcmdata/dclist.h
+++ b/dcmdata/include/dcmtk/dcmdata/dclist.h
@@ -22,52 +22,12 @@
 #ifndef DCLIST_H
 #define DCLIST_H
 
-#include "dcmtk/config/osconfig.h"    /* make sure OS specific configuration is included first */
-
-#include "dcmtk/ofstd/ofcast.h"
-#include "dcmtk/ofstd/oftypes.h"
-
 #include "dcmtk/dcmdata/dcobject.h"
+
+#include <deque>
 
 /// index indicating "end of list"
 const unsigned long DCM_EndOfListIndex = OFstatic_cast(unsigned long, -1L);
-
-/** helper class maintaining an entry in a DcmList double-linked list
- */
-class DCMTK_DCMDATA_EXPORT DcmListNode 
-{
-
-public:
-    /** constructor
-     *  @param obj object to be maintained by this list node
-     */
-    DcmListNode( DcmObject *obj );
-
-    /// destructor
-    ~DcmListNode();
-
-    /// return pointer to object maintained by this list node
-    inline DcmObject *value() { return objNodeValue; } 
-
-private:
-    friend class DcmList;
-
-    /// pointer to next node in double-linked list
-    DcmListNode *nextNode;
-
-    /// pointer to previous node in double-linked list
-    DcmListNode *prevNode;
-
-    /// pointer to DcmObject instance maintained by this list entry
-    DcmObject *objNodeValue;
-
-    /// private undefined copy constructor 
-    DcmListNode(const DcmListNode &);
-
-    /// private undefined copy assignment operator 
-    DcmListNode &operator=(const DcmListNode &);
-
-};
 
 /// list position indicator
 typedef enum
@@ -101,13 +61,13 @@ public:
     /// destructor
     ~DcmList();
 
-    /** insert object at end of list
+    /** insert object at end of list and make it be the "current" one
      *  @param obj pointer to object
      *  @return pointer to object
      */
     DcmObject *append(  DcmObject *obj );
 
-    /** insert object at start of list
+    /** insert object at start of list and make it be the "current" one
      *  @param obj pointer to object
      *  @return pointer to object
      */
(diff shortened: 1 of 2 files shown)


dcmdata/libsrc/dclist.cc
```diff
@@ -254,23 +165,11 @@ DcmObject *DcmList::seek( E_ListPos pos )
 
 DcmObject *DcmList::seek_to(unsigned long absolute_position)
 {
```
reunanen, line 167: Changing this O(n) search to an O(1) lookup is the beef of this PR.

michaelonken:
Hi,

thank you. This looks like an interesting patch (I did not try it yet); I have not been aware of std::deque which seems to be a good match for the given purpose. One "obvious" problem is that we still support compiling DCMTK with C++98 compilers (while using C++11 as the default). We might drop C++98 support in the future but right now we would need to make those guys happy as well. 

In the past we wrote our own classes (in ofstd, like OFVector, OFList, OFTuple etc.) as drop-in replacements for std lib features but that's probably too much work for std::deque also before the background that we might not need it any more in a few years. Another way to go is to only use the new behavior if C++11 or greater is available and otherwise fall back to the old  implementation.

What do you think?

reunanen:
Thank you for your answer. I think `std::deque` itself is included in C++98, but some minor parts of the PR (such as `std::next` and `std::prev`) are not. However, I think it should be possible to rewrite those parts so that they compile even on C++98.

So, I think two options here:
* Use the new behavior only if C++11 is available, as you suggest. I can do this, if you like.
* Update the new (`std::deque` based) code so that it should compile even on C++98. I could try this too, but I'm not sure how to test as I don't have a C++98-only compiler easily available I'm afraid.

dcmdata/include/dcmtk/dcmdata/dclist.h
```diff
@@ -101,13 +61,13 @@ class DCMTK_DCMDATA_EXPORT DcmList
     /// destructor
     ~DcmList();
 
-    /** insert object at end of list
+    /** insert object at end of list and make it be the "current" one
```
reunanen, line 64: Document the existing behavior for clarity (might be relevant on the call site).

jriesmeier:
I just read on [cppreference.com](https://en.cppreference.com/w/cpp/container/deque): "Insertion or removal of elements - linear O(n)."
This is where the existing double-linked list is probably better (in terms of complexity).

jriesmeier:
@michaelonken That's true, but reading DICOM files/datasets is not the only use case. I would rather suggest to check where @reunanen had performance problems with the existing implementation and try to enhance this particular use case.
For example, I can remember that parsing large DICOMDIR files or large Sequence of Items elements was pretty slow before we change the internal implementation (avoid iterating over an index).

reunanen:
Yes, that makes sense. I can try and see if I could easily change the higher-level implementation to use something that might be better suited for the task perhaps.

Unfortunately I don't have permission to share the example image, but I don't think it's extremely atypical or anything. For what it's worth: I started looking into this whole thing because a client (a hospital) mentioned that reading the exact same image pixels from a DICOM file was some 2x slower than reading in a corresponding [Aperio .svs](https://openslide.org/formats/aperio/) file.

The example file is a whole-slide pathology image that is about 1 gigabyte in size. While the image itself does not really look unusual to me, I'm not sure if the DICOM representation is "typical", at least just yet. (I'm certainly not a DICOM expert.)

jriesmeier:
I am aware of the DICOM representation of Whole Slide Microscopy Images (WSI). In fact, some of the performance improvements I referred to in my previous comment had been motivated by these huge WSI objects. Processing DICOM SOP Instances with 1 GBytes in size or even more than 100 GBytes in size is no problem at all if you know what you are doing (e.g., I implemented a software for this purpose based on the DCMTK for one of my customers). Iterating over the index of a data structure that is based on DcmList is certainly not a good idea, but I agree that this could (and should?) be improved.

michaelonken:
> @michaelonken That's true, but reading DICOM files/datasets is not the only use case. I would rather suggest to check where @reunanen had performance problems with the existing implementation and try to enhance this particular use case. For example, I can remember that parsing large DICOMDIR files or large Sequence of Items elements was pretty slow before we changed the internal implementation (avoid iterating over an index).

I know, however, I thought O(n) should also be the complexity for random insertion (or access in general) in DcmList.

reunanen:
I have an idea for an alternative implementation; let me try and see if I can get a similar performance improvement in my use case, without sacrificing the performance of inserting in the middle of the list (which is indeed O(1) at the moment).

I'll get back to you guys when I've been able to try it out.

reunanen:
Closing in favor of #114.

PR #115, opened 2025-02-05
Fix a documentation copy/paste error

malaterre:


jriesmeier:
Thank you for your report. This has just been fixed in commit 07dc3090a, which will soon be visible in the public Git repository.

PR #116, opened 2025-02-11
Provide a hook to setup iconv path at runtime

malaterre:


oficonv/libsrc/citrus_bcs.c
```diff
@@ -25,12 +25,21 @@
...
 
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
 
+char oficonv_buffer[4096];
+char *oficonv_path = NULL;
+
+void OFiconv_setpath(const char *buf) {
+  strcpy(oficonv_buffer, buf);
```
michaelonken, line 39: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

It appears that you are using `strcpy` or `strncpy` functions that does not affirm the size of the destination array and does not automatically NULL-terminate strings leading to buffer overflows risks. We recommend you to apply bound checking or null terminator. Learn more - https://www.geeksforgeeks.org/why-strcpy-and-strncpy-are-not-safe-to-use/

michaelonken, line 39: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

We observed that your code contains out-of-bounds read which is a type of memory access error that occurs when a program reads data from a memory address outside of the bounds of a buffer. This can result in the program reading data that does not belong to it, which can cause crashes, incorrect behaviour, or even security vulnerabilities. To prevent this you can perform bounds checking before accessing an array or other data structure. Learn more - https://www.martellosecurity.com/kb/mitre/cwe/125/

michaelonken, line 39: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

We observed your code contains an out-of-bounds write which may lead to security issues such as crashing, information leaks, or even arbitrary code execution. You should perform bounds checking, validate input sizes and indices to ensure they stay within the allocated memory bounds. Learn more - https://github.com/harsh-bothra/SecurityExplained/blob/main/resources/cwe-787.md

eichelberg:
> @eichelberg This is mostly a WIP but basically I need this API to specify where to search for esbd folders. Let me know if you want more details about the use case. Thanks !

Is there a reason why you cannot use the DCMICONVPATH environment variable to point to the desired directory?

eichelberg:
> getenv operates only on the data structures accessible to the run-time library and not on the environment "segment" created for the process by the operating system.

That means that you cannot call SetEnvironmentVariable() to change an environment variable in the running process and then expect the next call to getenv() to report that new value.

> You still have an issue when more than **one** path is used on a single system (setenv/getenv is fundamentally non-thread safe).

It is thread unsafe, but I am fairly certain that each process has its own set of environment variables, on Windows and on Posix systems. Another application calling setenv() cannot interfere with your application.

Anyway, I am not against adding a function in the oficonv module that allows you to set the iconv path manually.
Once the patch is complete and does not do stupid things like assuming that a path will never be longer than 4096 bytes, please let me know and I will happy to merge this into DCMTK.

eichelberg:
One more thing: DcmSpecificCharacterSet::setIconvPath() should return an OFCondition, and should actually return an error when DCMTK has been compiled with a different conversion library (GNU libiconv or libc iconv) that does not support such API.

jriesmeier:
By the way, I don't see a good reason why OFCharacterEncoding and DcmSpecificCharacterSet should be "polluted" with a static function that has nothing to do with these classes.

eichelberg:
This looks good now. I just merged this into our testing branch. It should appear in the public repository in a few days.

PR #117, opened 2025-02-11
Document case where FD with 1-N is partially read

malaterre:


jriesmeier:
@malaterre Your statement "For VM 1-N element only the first element is read from the file." does not seem to be correct. What makes you say that?

PR #120, opened 2025-06-13
Fixed compilation error

malaterre:
dcmtk/dcmdata/include/dcmtk/dcmdata/libi2d/i2d.h:37:24: error: ‘class Image2Dcm’ has virtual functions and accessible non-virtual destructor [-Werror=non-virtual-dtor]
   37 | class DCMTK_I2D_EXPORT Image2Dcm
      |                        ^~~~~~~~~


michaelonken:
Thanks Mathieu, closed with a96783acb19aafadbd3cbd96287da946494ad09c (visible on public master within the next few days).

PR #123, opened 2025-06-27
Properly null terminate string

malaterre:
This will prevent garbled output on stdout


eichelberg:
Merged into DCMTK's testing branch.

PR #126, opened 2025-08-22
Added support for Patient's Age in DSRDocument.

parasite-lost:
Added support for the optional attribute Patient's Age from the Patient Study Module. The new attribute is written to and read from DICOM dataset as well as to/from the dcmsr-specific XML document format.

Also extended test case to check whether the get and set methods work as expected.


jriesmeier:
@parasite-lost Thank you for your PR, which was almost perfect (only the definition of type dsr:AgeString was missing in the XML Schema). I've just committed the required changes to the local testing branch and with make the changes publicly available as soon as our test servers are happy with it.

PR #144, opened 2026-04-03
Make it easier to add external DCMTK modules

lassoan:
Added DCMTK_EXTERNAL_MODULES CMake variable to allow adding external modules to DCMTK without having to copy files into the DCMTK source tree.

DCMTK_EXTERNAL_MODULES contains a list of absolute paths, each pointing to a directory that follows the standard DCMTK module layout: it must contain a "CMakeLists.txt" at its root and an "include" subdirectory. Each module is added to the build via add_subdirectory() and its include directory is appended to DCMTK_INCLUDE_DIRS, both in the build tree and in the generated DCMTKConfig.cmake. Unlike DCMTK_MODULES, these paths are not relative to the DCMTK source tree, so the module source can reside anywhere on the filesystem.

This is used by [DCMTKcs](https://github.com/lassoan/DCMTKcs) - repository containing community supported modules for DCMTK. Currently, the only module is dcmjp2kcs module for JPEG2000 encoding/decoding.

### DIFF
--- a/CMake/GenerateCMakeExports.cmake
+++ b/CMake/GenerateCMakeExports.cmake
@@ -55,6 +55,9 @@ set(DCMTK_CONFIG_CODE "${DCMTK_CONFIG_CODE}list(APPEND DCMTK_INCLUDE_DIRS \"${DC
 foreach(module ${DCMTK_MODULES})
     set(DCMTK_CONFIG_CODE "${DCMTK_CONFIG_CODE}list(APPEND DCMTK_INCLUDE_DIRS \"${DCMTK_SOURCE_DIR}/${module}/include\")\n")
 endforeach()
+foreach(module ${DCMTK_EXTERNAL_MODULES})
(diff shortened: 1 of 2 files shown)


michaelonken:
I don't see a problem with this approach, also, because it leaves things within DCMTK as is and seems to be an elegant way to add extra modules, community-based or "private" within companies. 

However, maybe one of my DCMTK team colleagues wants to add something or has a different opinion.

jriesmeier:
I like it too, and I find it useful, but I'm no CMake expert.

lassoan:
Thanks for the reviews. The changes are very limited and have no effect if the optional `DCMTK_EXTERNAL_MODULES` variable is not explicitly set, so there is no risk in integrating them.

@eichelberg please have a look and merge if it looks good to you. It is also fine if you suggest an alternative way to add custom modules outside of the source tree (for example, the mechanism that you use already to integrate commercial modules).

michaelonken:
Merged in commit d974b87ca, visible upstream in a few days after running through our internal build dashboard.

lassoan:
Thank you @michaelonken for the review and merge.

PR #145, opened 2026-04-03
Fix OpenJPEG target handling in DCMTKConfig.cmake

lassoan:
When DCMTK is built with OpenJPEG support, DCMTKTargets.cmake contains
INTERFACE_LINK_LIBRARIES entries that reference the imported target
openjp2. If a consuming project includes DCMTKConfig.cmake without
previously finding OpenJPEG, CMake errors out because the openjp2
target does not exist.

Use CMakeFindDependencyMacro to conditionally call find_dependency(OpenJPEG)
when DCMTK_WITH_OPENJPEG is enabled. This guarantees that the openjp2
target is defined before DCMTKTargets.cmake is processed,
while avoiding redundant lookups in projects that have already found
OpenJPEG themselves.

### DIFF
--- a/CMake/DCMTKConfig.cmake.in
+++ b/CMake/DCMTKConfig.cmake.in
@@ -109,6 +109,11 @@ SET_AND_CHECK(DCMTK_TARGETS "@PACKAGE_DCMTK_CMKDIR_CONFIG@/DCMTKTargets.cmake")
 # Compatibility: This variable is deprecated
 set(DCMTK_INCLUDE_DIR ${DCMTK_INCLUDE_DIRS})
 
+include(CMakeFindDependencyMacro)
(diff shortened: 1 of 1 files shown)


michaelonken:
@eichelberg This is probably something for you.

lassoan:
@eichelberg please have a look. Without this, downstream projects need to add find_package calls depending on what DCMTK modules are enabled.

Note that the same problem is fixed for OpenSSL and zlib packages here: https://github.com/DCMTK/dcmtk/pull/67

eichelberg:
I have generalized this code a bit to also cover the other external libraries used by DCMTK, and committed the result as commit #c7359f8de. This will appear in the public repository once it has passed out nightly builds.

lassoan:
Awesome, thank you very much @eichelberg.

PR #147, opened 2026-04-16
Remove all invalid chars from UI

malaterre:


malaterre:
Original use case was a trailing nbsp (A0) in latin1 dataset

jriesmeier:
Thank you for your proposal, but I don't think it makes sense to remove all invalid characters/bytes from a UI value. Removing space characters (or all whitespaces) seems reasonable, as these can be found quite often in UI values, e.g., because the implementer did not read the VR definition in DICOM PS3.5 very carefully.

malaterre:
ACK. I'll re-consider your suggestion to turn into VR:UN any attribute that cannot be converted into JSON/UTf-8 now.

PR #148, opened 2026-05-04
Fix undefined behavior: shift of negative char in DcmHashDict::hash

MarkLee131:
When privCreator contains a byte with the high bit set, *privCreator (a plain char, signed on most targets) sign-extends to a negative int, and shifting a negative int is undefined behavior per the C/C++ standard. Cast through unsigned char first so the shift is well defined and the per-byte XOR pattern matches the original intent.

Same class of fix as 519187efe (2016, Riesmeier).

Detected by UBSan during fuzz testing.


michaelonken:
Thank your the report and fix. I integrated it into our internal testing branch as commit 6043bf669 whic will be visible on public master within a few days.

PR #154, opened 2026-07-14
COMP: Use single export() call to fix namespace corruption

blowekamp:
Replacing three APPEND export() calls with one non-APPEND call fixes a namespace corruption bug where CMake prepends the DCMTK:: namespace to dependency targets (e.g. ITK codec targets) that belong to other export sets.

CMake's HandleMissingTarget only searches other registered export sets when not in append mode. In append mode it blindly prepends the current namespace, producing DCMTK::ITK::ITKZLIBModule instead of the correct ITK::ITKZLIBModule in DCMTKTargets.cmake.

The file(WRITE)+3xAPPEND pattern was used only to accumulate targets across multiple calls; combining them into a single export() call is equivalent and avoids the append-mode shortcut.


dzenanz (approved):

hjmjohnson (approved):
Thanks @blowekamp Reviewed, built local, and confirmed this resolved the linking issues downstream.

PR #156, opened 2026-08-11
Fixed TrcTrackSet::getTrackSetNumber

ppmartel:
Discovered that TrcTrackSet::getTrackSetNumber was always returning an error. Changed it to use Uint32 instead of Uint16, as the TrackSetNumber tag has a VR of UL. Local tests showed this fix to now properly return the value.


michaelonken:
Hi Phil!

Thank you very much for the report and patch. I just merged it into our testing branch and the commit will be visible within th next days on public master branch.

BR Michael

Fixed with commit 0e9fd382f.

PR #158, opened 2026-09-22
Make sure to handle the case no overlay and return '2'

malaterre:


malaterre:
@eichelberg Here is what the doc states:

 

>     ** @return false (0) if an error occurred, true otherwise (1 = planes have been successfully deactivated,
>      *                                                          2 = no planes to be deactivated)
>

jriesmeier:
@malaterre Not sure why you reference "eichelberg": The dcmimgle/dcmimage has been developed by my younger self (a few decades ago) ;-)

Nevertheless, the return value 2 is only used for monochrome images, as the DicomImage class only supports overlay planes for monochrome images. For color images, 0 should always be returned.

malaterre:
> @malaterre Not sure why you reference "eichelberg": The dcmimgle/dcmimage has been developed by my younger self (a few decades ago) ;-)

Well git log seems to indicate Marco ...

https://git.dcmtk.org/?p=dcmtk.git;a=history;f=dcmimage/include/dcmimage.h

> Nevertheless, the return value 2 is only used for monochrome images, as the DicomImage class only supports overlay planes for monochrome images. For color images, 0 should always be returned.

My comment was about the '2' returns value. As a user I want to write:

```
  const int hide = scaled->hideAllOverlays();
  if (hide==0) {
    std::cerr << "Cannot hide all overlays";
    return false;
  }
```

Let me know if the above is proper understanding of doc and return value 1/2

jriesmeier:
> Well git log seems to indicate Marco ...

Yes, he did the first commit, but the original source code was developed as part of my diploma thesis. The main/original author of each file is indicated in the comment header at the beginning of each file.

> Let me know if the above is proper understanding of doc and return value 1/2

The basic idea (at that time) was: 0 indicates an error, everything else success. This was before the OFCondition class was introduced. For some methods/functions, the success status distinguishes between different types of success. In this case, 1 means "[all] planes have been successfully deactivated" and 2 means "[there were] no planes to be deactivated".

malaterre:
@jriesmeier I believe this is the part that caused my confusion, this is not clear where you got it from the doc:

>  [...]For color images, 0 should always be returned.

So user code should be instead:

```
  const int hide = scaled->hideAllOverlays();
  // samplesPerPixel == 3 is always an error
  if (hide==0 && samplesPerPixel==1) {
    std::cerr << "Cannot hide all overlays";
    return false;
  }
```

thanks (again) !

jriesmeier:
@malaterre I would suggest something like the following:

```
if (scaled->isMonochrome())
{
  if (!scaled->hideAllOverlays())
  {
    // error...
    return false;
  }
}
```

